PR #2, opened 2026-08-11
Revise security policy for issue reporting

callghost:
Updated the security policy to focus on reporting security issues through GitHub advisory and removed details about supported versions and reporting vulnerabilities.

PR #8, opened 2026-08-14
Add Agent Orchestration

rvanhorn:
Hello,

Thanks for building VelaTerm and making it open source! It checked almost every box I was looking for except agent orchestration, so I decided to add it myself. :)

## Summary

This PR adds multi-agent orchestration to VelaTerm.

A lead agent can now split up work, send tasks to child agents, monitor them, review their changes, and merge completed work.

Highlights:

* Child agents can use different agents, models, and effort levels.
* Parent/child session relationships are persisted.
* Workers can optionally use isolated Git worktrees.
* New `vagent` commands manage workers, results, diffs, merges, and cleanup.
* New `/orch` skill handles autonomous delegation and supervision.
* Reusable routing profiles can define agent, model, effort, permissions, and worktree settings.
* Backend limits control worker count, parallelism, nesting depth, confirmations, and wait timeouts.
* Worker failures include provider error details when available.

## How It Works

`vspawn` and `vspawn-tree` create child sessions under the current session. VelaTerm validates the request, applies orchestration limits, optionally creates a Git worktree, saves the worker settings, and launches the selected agent with the correct CLI arguments.

The new `vagent` command provides:

* `spawn`
* `spawn-status`
* `config`
* `list`
* `status`
* `wait`
* `read`
* `prompt`
* `cancel`
* `diff`
* `merge`
* `cleanup`

Sessions can only manage their own descendants.

The `/orch` skill turns the current agent into a lead agent that can split up work, choose configured routing profiles, launch workers, wait for results, review changes, send follow-up prompts, and merge finished work.

`/vspawn` is still available for simple manual delegation.

## Git Integration

Workers can use separate Git worktrees for parallel changes.

* `vagent diff` reviews worker changes.
* `vagent merge` merges a worker branch.
* `vagent cleanup` removes completed worktrees.

The existing graphical Git workflow is unchanged.

## Known Limitations

* Permission behavior still uses VelaTerm's existing permission system.
* Failure details depend on transcript support. Claude, Codex, and Grok are supported; other agents may report `unknown`.
* The orchestration settings modal does not yet have automated visual tests.

## Testing

Passed:

* `pnpm test` — 203 tests across 27 files.
* `cargo test --manifest-path src-tauri/Cargo.toml` — 408 passed, 8 ignored because they require external tools, credentials, or local agent history.
* `pnpm build`
* Targeted ESLint checks on changed TypeScript files.
* `git diff --check`

Tests cover worker failures, provider errors, orchestration limits, routing profiles, agent launch arguments, and Git worktree creation/diff/merge/cleanup.

## Breaking Changes

None intended. Existing `vspawn`, `vspawn-tree`, session persistence, and permission behavior remain compatible.


mguttmann:
Really nice work — I read through the code and would like to raise one design point before this merges: **the worker session lifecycle has no end state**, and the spawn limits interact with that in a way that caps larger orchestration runs.

## What the code does today

1. **Finished workers stay in the session tree forever.** `cancel` deliberately keeps the session alive, and `cleanup` only removes the worktree and the disposable branch — `op_cleanup` (`src-tauri/src/agent/ctl.rs`) clears the session's worktree link via `clear_node_worktree`, but never archives or removes the session itself. The `vagent` op set (`list / status / wait / read / prompt / cancel / diff / land / spawn / spawn-status / config / cleanup`) has no operation to retire a finished child.

2. **At the same time, `max_descendants` counts *retained* descendants.** `subtree_counts()` counts every descendant regardless of state, and the spawn guard says so explicitly: *"Waiting does not free a slot; archive or remove a finished session."* With the default of 10, a lead hard-stops (HTTP 429) after 10 spawns until the human manually deletes sessions in the left panel.

## Why this matters

Agent workflows routinely burn through dozens of short-lived workers per task — a single feature run can easily use 50–100. With the current design the human becomes the garbage collector: either the tree fills up with dead sessions (limit raised), or the lead stalls at 10 (default). Neither is great for autonomous orchestration.

## Suggestion

Give the lifecycle an explicit end that the lead itself can trigger, for example:

- a `vagent archive <id|name>` op, guarded like `cleanup` (only exited children, only with a verified landing), and/or
- an opt-in on `cleanup --confirm` (or a setting like `retainFinishedSessions: false`) that retires the session together with its worktree once its landing is verified.

Retain-by-default is a fine philosophy — being able to reopen a worker later is genuinely useful. But the lead needs *some* sanctioned way to complete a full `spawn → land → cleanup → retire` cycle, otherwise `max_descendants` turns from a safety limit into a hard cap on total work per lead.

Happy to test a revision — I run large multi-agent workflows daily, so this is exactly the scenario I would hit.

rvanhorn:
Good catch, thanks for review!

 I missed that finished descendants would make `maxDescendants` a hard cap. I've added a new `vagent` flow to deal with that.

mguttmann:
That was fast — thank you! I verified the three follow-up commits against the concern I raised, and the core issue is solved: `vagent retire` archives the settled subtree recursively, every descendant-facing op filters on `live_descendants` (`archived_at IS NULL`), and your test `retire_archives_a_settled_worker_and_frees_its_descendant_slot` proves the exact scenario end to end (429 on `max_descendants` → `retire --confirm` → the next spawn succeeds). The guards reuse `verify_cleanup`, retirement lands in the existing restorable ArchivePanel, and CLI/docs parity is complete. Nice work.

While verifying I read the new code closely and found a few things — two I'd fix before merge, the rest batched as follow-up material.

## 1. `retire --confirm` is not atomic; one mid-loop git failure wedges the subtree against every vagent op

The confirm path runs per worker: `worktree_remove` → `branch_delete` → `clear_node_worktree` (`ctl.rs:862-879`), archives only at the very end (`:880`), and — unlike `op_cleanup`, which collects per-worker errors into a `failed[]` array and keeps going (`:714-731`) — aborts the whole op with 500 on the **first** error.

Worst case: `worktree_remove` succeeds (`:863`) but `branch_delete` fails (`:866`). The early return skips `clear_node_worktree`, so the session keeps a `worktree_path` pointing at a removed directory. Every retry then re-runs `verify_cleanup` against that missing path, which yields `CleanupCheck::Blocked` (`:635-636`), so `verify_retire_worktrees` hard-fails 409 (`:762-776`) on every subsequent attempt. Processes are already killed, nothing is archived, and nothing in the vagent API recovers the subtree — `cleanup` lists it as blocked, retire keeps 409ing, and retiring the parent from *its* parent hits the same verification. The only ways out are the unguarded UI archive from point 3 (which leaks the branch) or manual git repair.

Suggestion: give retire the same per-item `failed[]`-and-continue semantics as cleanup, or make it resumable — e.g. treat a recorded landing whose worktree directory is already gone as "cleanup done, continue to archive". Related ordering nit: the kill loop (`:852-856`) runs *before* the second `verify_retire_worktrees` (`:857-860`), so a 409 on the re-verify has already terminated every shell in the subtree without archiving anything. Swapping the two would make that failure path side-effect-free — at the price of verifying before the processes are quiesced, a trade-off that admittedly matters less than it looks given that the kill is fire-and-forget anyway (it returns without waiting for process exit, so the current order doesn't guarantee quiescence either — a short wait-for-exit would fix both at once).

## 2. The effort-options change in 42978fe leaves the branch's own parity test failing — and its runtime effect quietly defeats `autoApprove`

`SpawnConfirmModal.tsx` now offers claude `[low, medium, high, xhigh, max]` and codex `[minimal, low, medium, high, xhigh]`, but the follow-up contains no matching `inject.rs` change. Your own test suite already flags this: `known_launch_values_match_the_spawn_card` (`ctl.rs:2841-2871`) parses the modal via `include_str!` and asserts `EFFORT_OPTIONS` equals `inject::known_efforts` **exactly**, and `codex_launch_values_use_current_chatgpt_identifiers` (`:2874-2883`) pins codex efforts to `[low, medium, high, xhigh, max]` — so `cargo test` should now be red for both agents (claude gained `xhigh`/`max`, codex gained `minimal` and lost `max`, which SKILL.md line 52 still advertises).

The runtime effect if the UI list shipped as-is: a curated-values warning forces the confirmation card and disables `autoApprove` for that spawn (`force_confirm = … || !launch_warnings.is_empty()`, `ctl.rs:461-462`; the claude+`xhigh` warning is itself asserted at `:2657-2659`). So picking one of the newly offered values from VelaTerm's own dropdown puts a human back into every spawn that uses it — a lead could suppress that with `--allow-unknown-launch-values`, but it has no reason to expect values from the app's own dropdown to need it. Syncing `known_efforts` with the intended UI lists (and SKILL.md's codex line) resolves all three at once.

## 3. Class issue: the unguarded UI archive path vs. the guarded retire path

The pre-existing one-click "Archive Session" in the tree menu calls `set_session_archived` directly — no settled check, no landing verification, no worktree cleanup. Now that archiving is the sanctioned lifecycle end state, this collides with the retire guards in both directions:

- **Archive-side:** a worker archived from the UI vanishes from `live_descendants`, so `cleanup` (`ctl.rs:670-676`) and `retire` (404 from `resolve_named` at `:1336-1339`) cannot see or reclaim its worktree and branch while it stays archived — they leak, invisible to every vagent op, until someone happens to restore the session. Meanwhile the `max_descendants` slot frees *without* any verification: the guarded path blocks on unlanded work, the unguarded path archives it in one click.
- **Restore-side:** restoring a retired worker from the ArchivePanel silently makes it count against `max_descendants` again (a lead can hit 429 through no action of its own), and the archive contract "data and recordings remain, allowing exact agent resume after restoration" (`repo.rs:1716-1719`) no longer holds — retire already deleted its worktree and branch.

Suggestion: route the UI archive of a session that still holds a `worktree_path` through the same verification/cleanup chokepoint (or at minimum warn), and document restore semantics for retired workers.

## 4. For the large-run scenario, the human moved rather than disappeared

All four default profiles set `worktree: true` (`orchestration.rs:191, 194-231`), so nearly every retire previews as `cleanup-and-archive` — and SKILL.md step 16 requires showing the user the deletion list per worker. On the 50–100-worker runs from my original comment, that's one approval per worker: the human went from "delete sessions in the left panel" to "approve retire lists". Meanwhile the rule is skill-text-only — `op_retire` checks a plain request bool (`ctl.rs:818`), while spawn has a real backend-enforced card (`:461-476`); the skill even contradicts itself (line 19: "No per-action approval is needed beyond VelaTerm's own spawn confirmation card" vs. step 16). Two ways out, either works: a backend retire-confirmation card with an `autoApprove`-style setting (consistent with spawn), or document the batched route — `vagent cleanup --confirm` first (one approval), after which each retire previews as plain `archive` and runs autonomously (for flat trees, at least — cleanup only covers direct children, so a nested worktree worker still flips its subtree's preview back to `cleanup-and-archive`).

<details>
<summary><b>Smaller observations (batched)</b></summary>

- **Settled check trusts the laggy state cache** (`ctl.rs:802-816` via `state_of`): SKILL.md itself documents the lag; a child whose new turn hasn't hit the cache yet passes as "waiting" and gets killed mid-turn. Non-worktree children get no post-kill dirty check at all.
- **`PtyManager::kill` is fire-and-forget** (returns without waiting for process exit — the error branch at `ctl.rs:853-855` is dead code since it always returns Ok): a dying agent can write into its worktree after the cleanliness check, failing the non-force `worktree remove` mid-loop and triggering the wedge from point 1.
- **Retired = nonexistent to the API**: after a successful retire, a second retire (and `read`/`status` on the worker) returns the same 404 as a typo'd name. An explicit "session is archived" response would let a lead distinguish "already done, proceed" from "investigate".
- **Preview is fail-fast, skill promises aggregation**: SKILL.md step 15 says "Report every blocked worker and its reason", but both the settled loop (`ctl.rs:802-816`) and `verify_retire_worktrees` (`:742-780`) return only the *first* blocker — even in preview mode. `cleanup` aggregates `blocked[]`/`failed[]`; retire's preview should too (or the skill wording softened).
- **"Dirty or unlanded worktrees block retirement" is worktree-only**: plain (non-`--worktree`) children are archived with zero work verification (`ctl.rs:748-754` skips them). Defensible — nothing is deleted — but worth one explicit sentence in the skill so a lead doesn't over-trust the guard.
- **Growth moved, not removed** (fine for merge, follow-up idea): retired workers accumulate in the ArchivePanel with their recordings; the panel's existing delete is per-session, so at orchestration scale a bulk "delete retired workers older than N" action would close the loop.

</details>

None of this blocks the core fix — the hard cap is demonstrably gone and the lifecycle design (guarded, two-phase, reversible via archive) is sound. Points 1 and 2 are what I'd want addressed before merge; 3 and 4 could be follow-up issues if the maintainers prefer. Happy to re-verify a revision.

mguttmann:
Re-verified as promised — I pulled the branch, read the revision closely, and ran both suites on a fresh clone: **`cargo test` 503 passed / 0 failed** (the `known_launch_values_match_the_spawn_card` parity test is green again), **`pnpm test` 283 passed**, **`pnpm build` green** (which, via the `Dict = typeof en` typecheck, also proves all 11 locales carry the new retire strings).

Every point from my comment is addressed, and mostly beyond what I asked:

- **P1 non-atomicity**: both suggested options delivered — per-item `failed[]`-and-continue with archive skipped while anything failed, *and* resumability (`RetireCleanup::Resumed` recovers a missing-directory worker from its landing record instead of blocking forever), with regression tests for the partial-cleanup resume and the never-landed block.
- **P1 kill ordering / fire-and-forget**: the post-card recheck now runs *before* the kill loop (with a plan-equality 409 if anything changed), and `kill_and_wait` + `procstat` actually waits for the process tree to exit, blocking with a clear message on timeout.
- **P2**: effort lists re-synced (UI aligned back to the backend's curated list) — proven by the green parity test.
- **P3**: the UI archive now goes through `prepare_archive`, which reuses the retire verification, refuses running owners/descendants, and warns with the deletion list at all three entry points; restore-exactness is documented in the ArchivePanel banner, repo docs and the manual.
- **P4**: a real backend retire-confirmation card (`retire://request`, answered via `retire_result`) plus an `autoApproveRetire` setting (default false, and resumed rows force the card regardless — nice touch), the skill self-contradiction fixed, and the batched cleanup-first route documented.
- **M1/M3/M4/M5**: settle verdict now uses cache-independent signals (transcript quiescence + input-revision-pinned turn completion — `retire_refuses_a_low_cpu_turn_started_after_cached_waiting` covers exactly the lag hole), retired workers answer an explicit 409 `archived` instead of a 404, the preview aggregates every blocker, and the worktree-only scope of the guard is spelled out in the skill.

Two new things surfaced while verifying the revision itself — the first one matters:

## 1. The resume path force-deletes a branch on the strength of an *unverified* landing row

`resume_retire_cleanup` treats "a landing row exists" as "cleanup already verified": it only checks that `get_agent_landing` returns `Some` and reads `source_branch` (`ctl.rs:804-813`) — it never checks `landing.target_commit`, the parent match, or that the branch tip is on the target branch. But a landing row is persisted *before* verification: `begin_agent_landing` inserts the row with `target_commit = NULL`, and its `ON CONFLICT` arm even resets a previously **verified** row back to NULL (`repo.rs:1100-1132`); `op_land`'s non-conflict error paths return 500 and leave that unverified row in place. `verify_cleanup` itself refuses exactly these rows ("worktree has no verified landing", `ctl.rs:673-676`) — the resume path accepts them.

Scenario: worker lands successfully and keeps working; a later re-land fails after `begin_agent_landing` (row now unverified); the worktree directory later disappears (crash, manual delete). Retire then plans the worker as `resumed`, and `retire_one` runs `branch_delete_if_present` → `git branch -D` (`git.rs:1190-1202`, force) — destroying the unlanded commits. The retire card does fire for resumed rows, but its premise ("only the branch remains to delete") is what's wrong here.

Fix is small: hold the resume path to the same bar as `verify_cleanup` — require `target_commit` to be set (and ideally the `is_ancestor` check) before force-deleting; otherwise block like the never-landed case already does (`ctl.rs:809-811`).

## 2. The new archive chokepoint regresses a pre-existing plain-UI flow

`prepare_archive` now owns **every** parented session holding a worktree — including sessions created manually via "New Child Session → New Agent Session → worktree" (a flow that exists on main, outside orchestration). Those sessions normally have no vagent landing record (one only exists if the parent is itself an agent session and ran `vagent land` on the child), so a one-click archive that worked before the PR is now refused with no in-dialog path forward: "Land or retire the worker first." (`ctl.rs:1190-1200`) — vagent concepts a plain UI user doesn't have. The escape hatch (Git → Delete worktree…, then Convert to normal session, then archive) exists but is undiscoverable from the error. Suggest either scoping the guard to sessions with orchestration provenance (a landing record, or spawned via vagent), or teaching the ArchiveBlocked dialog the actual path ("delete the worktree first, then archive"). To be fair: this is the trade-off my own P3 ask invited — it just needs a deliberate answer for the non-orchestration case.

<details>
<summary><b>Smaller observations (batched — none blocking)</b></summary>

- **Residual dirty-after-kill state**: two post-kill paths still return 409 with every shell dead and nothing archived (exit timeout, post-kill rescan). If an agent wrote into its *verified* worktree in that window, every retry blocks on "worktree has uncommitted changes" and no vagent op can commit/discard for the killed worker — same class as the old wedge, now visible and confined to retire retries, but recovery is still manual git. Related: `kill_and_wait` snapshots `subtree_pids` before killing, so a process spawned in between escapes the wait.
- **`prepare_archive`'s running-process guard scans only the archived subtree** (`ctl.rs:1138-1141`): a running session *outside* it whose cwd is inside an owned worktree — most plausibly the lead cd'd into a child worktree to inspect it — is never consulted, and the directory is removed from under it, contradicting the guard's own comment at `:1109`. Checking every running PTY's cwd would close it.
- **Unclamped retire card timeout**: `confirm_retire` takes `timeoutSecs` verbatim (`ctl.rs:1397-1401`); `--timeout` near `u64::MAX` overflows into an untimed `recv`, parking the handler thread forever with the card open. Separately, `RETIRE_WAITERS` is pruned only when the *next* waiter registers (TTL 3600s), so an unrelated retire can expire a long-parked card early. Clamping `timeoutSecs` to `OUTCOME_TTL` covers both.
- **A single stray keystroke into a waiting worker's pane permanently invalidates the turn-completion signal** (`manager.rs:1239-1242` bumps the revision; `:219-236` requires equality): retire then refuses a running idle worker forever, and the skill's "wait, then run it again" never resolves it. Fails in the safe direction, but the recovery (prompt the worker into a new turn, or `vagent cancel`) deserves a skill sentence — or reset the revision on the next authoritative waiting event.
- **UI archive: confirmation and deletion live on opposite sides of the IPC boundary.** The ConfirmArchive dialog is derived from frontend state, while the backend deletes verified worktrees without further confirmation once called — a stale view archives with worktree deletion and *no* card; conversely the dialog can list paths the backend deliberately keeps (frontend `archivePlan.ts` excludes only group bindings, backend also excludes session-shared ones). Two adjacent nits: the card's text "The work is already on the parent branch." is asserted before any verification has run (false for unlanded workers — soften to "only verified landed work is deleted"), and two subtree sessions binding the same path produce two plans for one directory (first removes it, second fails, `prepare_archive` aborts half-done via `?` at `ctl.rs:1183-1185`). Returning the backend-computed plan — as the retire card already does — would resolve the class.
- **`agent_available` probes run `$SHELL -lc`/`-lic` with no timeout inside the dispatch/config path** (`install.rs:368-401`): a blocking rc file hangs `orchestration_default_profiles` (and, on a fresh install, the first `vagent config`), and the `OnceLock` caches the hang for the process lifetime. A kill-after-N-seconds wrapper would bound it.
- **`retire_result` is deliberately remote-answerable** (forwarded to LAN clients, not in the gate lists) — consistent with the spawn-card precedent and the stated threat model, so it looks intended; but the gating comment's rationale is "no remote UI calls them", and here remote UI *does* render the card. One sentence in the `dispatch.rs` comment (or a pin test) recording the decision would keep the ACL story honest. Optional: broadcast `retire://cancel` once a card is answered, so other connected clients don't keep a stale card open.
- **Docs drift (pre-existing, same class the revision otherwise cleaned up)**: SKILL.md still documents a `merge` command with `refs/vlx/presquash`/`backupRef` that doesn't exist in the ops or in git.rs (`SKILL.md:110-114`), the "Land one commit for each child" section is empty (`:167`), and the vagent CLI usage text still describes pre-revision retire semantics. Also still undocumented: restoring a retired worker from the ArchivePanel silently counts it against `maxDescendants` again.

</details>

Bottom line: all ten previous points are demonstrably fixed and the whole thing runs green — this went from "orchestration demo" to a lifecycle I'd trust. #1 above is the one thing I'd want decided before merge (it's the old wedge class made destructive); #2 is a product call about non-orchestration users; everything else is follow-up material. Happy to take another pass after the next push.

mguttmann:
Since this merged into `feat/agent-orchestration` rather than `main`, I ran a full pre-merge audit of the whole feature against the branch head (`77eb22c`), including the surfaces my earlier comments did not cover: the spawn path, the hook server, the transports and the persistence layer. Both suites are green at that head (cargo 503/0, vitest 283, build ok), so everything below is behavioural rather than a test failure.

Filed so the findings do not get lost in a closed thread:

- #34 `land` retry can silently discard the user's staged work in the parent checkout (`git reset --merge` on a retry that cannot tell whose dirt it is)
- #35 a stale landing row permanently wedges `land` and `cleanup`, with no reset path in any op
- #36 `op_land` is not serialized, so concurrent lands race on the shared parent worktree
- #37 the unverified-landing branch force-delete from my last comment, now also reachable through the plain Archive action and the ungated WS command, without the card the retire path deliberately forces
- #38 `spawn://request` fans out to every connected client, so desktop plus a browser tab creates two sessions and two worktrees per request
- #39 the terminal renderer default changed from `dom` to `canvas` in this branch while the manual it ships still says DOM
- #40 collected smaller findings, including the items from my previous comment that are still open

Security-relevant findings from the same audit went through private vulnerability reporting as SECURITY.md asks, not into these issues.

The ten points from my previous review are all genuinely fixed, and the lifecycle work holds up well under close reading. Happy to re-verify any of the above once they are addressed.

PR #9, opened 2026-08-14
Fix AppImage failing to start from a desktop launcher (anchor WEBKIT_EXEC_PATH to the bundle)

metadan:
## Problem

The Linux AppImage fails to start when launched from a desktop entry. No window appears and the process aborts with:

```
** (velaterm:22177): ERROR **: Unable to spawn a new child process: Failed to spawn child process
?././/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitNetworkProcess? (No such file or directory)
```

## Cause

The bundled WebKitGTK resolves `WebKitNetworkProcess` and `WebKitWebProcess` through a path that is relative to the **working directory** (`./lib/<arch>/webkit2gtk-<abi>/`) rather than anchored to the bundle. The helpers actually ship at `$APPDIR/usr/lib/<arch>/webkit2gtk-<abi>/`, and `AppRun` exports `APPDIR`, `LD_LIBRARY_PATH` and `XDG_DATA_DIRS` but does not `chdir`, so the relative path is resolved against whatever directory the process was started in:

| Launched from | Working directory | Result |
| --- | --- | --- |
| Desktop launcher | `$HOME` or `/` | No match — startup aborts |
| Shell in `/usr` | `/usr` | Finds the **system** helpers |
| Shell elsewhere | varies | No match — startup aborts |

Note the middle row: where the host has `webkit2gtk` installed, the lookup silently pairs the system helper processes with the bundled WebKit library. That is why the failure looks intermittent rather than absolute, and why it can appear to work while running a mismatched pair.

Verified against the shipped `VelaTerm_0.1.100_amd64.AppImage` on Pop!_OS 24.04 (COSMIC/Wayland, XWayland). Setting `WEBKIT_EXEC_PATH` to the bundled helper directory makes it start correctly from any working directory, including `/`.

## Fix

Locate the helper directory under `$APPDIR` and export `WEBKIT_EXEC_PATH` before WebKit initializes, next to the existing `WEBKIT_DISABLE_DMABUF_RENDERER` workaround in `main.rs`:

- The lookup lives in `appimage.rs`, whose module doc already notes that WebKit's own processes *do* need the bundled paths — this makes that hold regardless of working directory.
- The architecture triple and the WebKit ABI version are discovered rather than hard-coded, so the lookup survives a cross-architecture build and a `webkit2gtk-4.1` → `4.x` bump.
- Applied only when the user has not already set `WEBKIT_EXEC_PATH`, matching the existing DMABUF override's behaviour.
- Outside an AppImage the lookup returns `None` and nothing changes; on non-Linux targets the call is compiled out.

`scripts/release.sh` (which writes the custom `AppRun`) is not in the public tree, so the equivalent one-line `export WEBKIT_EXEC_PATH=...` there was not an option from outside. Fixing it in `main.rs` also covers anyone running the binary directly from an extracted `AppDir`. If you would rather carry it in `AppRun`, this change can be dropped in favour of that.

## Tests

Three unit tests in `appimage.rs` cover the lookup: a bundle with helpers (using an architecture triple and ABI version deliberately unlike the current build, to prove neither is hard-coded), a bundle whose `webkit2gtk-*` directory holds no helper binary, and a root that is not a bundle at all.

Note on verification: the new code and its tests are compile-clean under `rustc --test -D warnings` and pass, but I was not able to build the full crate locally — the machine lacks `libwebkit2gtk-4.1-dev`/`libgtk-3-dev`, so `cargo test` cannot link the Tauri dependencies. Please let CI confirm the whole-crate build.

### DIFF
--- a/src-tauri/src/appimage.rs
+++ b/src-tauri/src/appimage.rs
@@ -105,6 +105,42 @@ fn overrides() -> Overrides {
     }
 }
 
+/// Directory holding the WebKitGTK helper processes shipped in this bundle, or `None` outside an
+/// AppImage and when the bundle does not carry them.
+///
+/// The bundled WebKit resolves `WebKitNetworkProcess` and `WebKitWebProcess` through a path that is
+/// relative to the working directory rather than anchored to the bundle, so it finds them only when the
+/// process happens to start from a directory that contains a matching `lib/<arch>/webkit2gtk-<abi>/`
+/// tree. Callers pass the result to `WEBKIT_EXEC_PATH` to make the lookup independent of that.
+pub fn webkit_exec_path() -> Option<std::path::PathBuf> {
+    find_webkit_exec_path(std::path::Path::new(appdir()?))
+}
+
+/// Scan one bundle root for the helper directory. Split out from process state so it is testable.
+///
+/// The architecture triple and the WebKit ABI version are discovered rather than hard-coded, so the
+/// lookup survives a cross-architecture build and a `webkit2gtk-4.1` to `4.x` bump.
+fn find_webkit_exec_path(appdir: &std::path::Path) -> Option<std::path::PathBuf> {
+    // $APPDIR/usr/lib/<arch-triple>/webkit2gtk-<abi>/WebKitNetworkProcess
+    for arch in std::fs::read_dir(appdir.join("usr").join("lib")).ok()?.flatten() {
+        let Ok(entries) = std::fs::read_dir(arch.path()) else {
+            continue;
+        };
+        for entry in entries.flatten() {
+            let dir = entry.path();
+            if dir
+                .file_name()
+                .and_then(|n| n.to_str())
+                .is_some_and(|n| n.starts_with("webkit2gtk-"))
+                && dir.join("WebKitNetworkProcess").is_file()
+            {
+                return Some(dir);
+            }
+        }
+    }
+    None
+}
+
 /// Value a variable should have in a child process, with AppImage paths removed. `None` means the child
 /// must not receive it at all. Use this instead of `std::env::var` when building a child's environment.
 pub fn clean_var(key: &str) -> Option<String> {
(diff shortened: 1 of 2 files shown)


metadan:
Closing this: the fix does not work. My verification was faulty and I would rather withdraw it than leave a wrong patch open.

**What I got wrong.** I tested the AppImage with `WEBKIT_EXEC_PATH` set and it started, so I concluded the variable was the fix. But those test runs also had the working directory set to `/`, and on this distribution `/lib` is a symlink to `usr/lib`. The relative lookup `././/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitNetworkProcess` therefore resolved to `/usr/lib/...` — the **system** helper. The app started because of the working directory, not because of the variable.

**Retested, isolating the two factors** (0.1.100 AppImage, Pop!_OS 24.04):

| Working directory | `WEBKIT_EXEC_PATH` | Result |
| --- | --- | --- |
| `$HOME` | unset | dies |
| `$HOME` | set to the bundled helper dir | **dies** — the variable has no effect |
| `/` | unset | starts, using the **system** helper via the `/lib` symlink |
| `$APPDIR/usr` | unset | starts, using the **bundled** helper |

So this build of WebKitGTK ignores `WEBKIT_EXEC_PATH` entirely. Only the working directory changes the outcome, and `$APPDIR/usr` is the one value that reaches the bundled helpers. Confirmed by checking `/proc/<pid>/exe` of the running `WebKitNetworkProcess`.

**The underlying bug is still real.** Launched from a desktop entry the working directory is `$HOME`, and the app aborts before a window appears:

```
** (velaterm:49996): ERROR **: Unable to spawn a new child process: Failed to spawn child process
“././/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitNetworkProcess” (No such file or directory)
```

Reproducible with `systemd-run --user --scope -- /path/to/VelaTerm.AppImage` from `$HOME`. On a host with `webkit2gtk-4.1` installed and a working directory of `/`, it starts against mismatched system helpers instead, which is why it can look intermittent.

**What a correct fix looks like.** Change the process's working directory to `$APPDIR/usr` before WebKit initializes — either a `cd "$HERE/usr"` in the `AppRun` that `scripts/release.sh` writes, or `set_current_dir` early in `main()`. If done in `main()` it needs care: `open_project_from_args` resolves relative paths against `std::env::current_dir()`, so the launch directory has to be captured before the chdir and passed through, otherwise `vela <relative-path>` breaks.

I did not want to guess a second time on your codebase's behalf. Happy to open a replacement PR along those lines if the `main()` approach is the one you want, or to leave it with you if `AppRun` is the better home — that file is not in the public tree, so you would be better placed to change it.

Apologies for the noise.

PR #10, opened 2026-08-14
Show the working tree's real changes in the Git tab (fixes #3)

### LINKED ISSUE #3: Git "Changes" panel never refreshes — contents stay frozen at first render
## Summary

The Git **Changes** panel renders once and then never updates. The file list, status letters and +/- line counts stay frozen regardless of what happens in the working tree, so the panel steadily drifts out of sync with the actual repo state.

## Environment

| Item | Value |
| --- | --- |
| VelaTerm | 0.1.100 |
| Commit | 7ccc7da72d3703860e26d3755b3a95f0bc0fa460 |
| Install | AppImage |
| OS | Pop!_OS 24.04 LTS |
| Kernel | 7.0.11-76070011-generic |
| Desktop | COSMIC (Wayland) |
| Backend | XWayland (AppRun forces `GDK_BACKEND=x11`) |

## Steps to reproduce

1. Open a session whose working directory is a git repo with uncommitted changes.
2. Observe the Changes panel populate.
3. Edit a tracked file, create a new file, `git add`, or `git commit`.
4. Look at the Changes panel again.

## Observed

The panel content is identical to step 2 — same files, same M/A letters, same +/- counts. It stays that way indefinitely:

```
CHANGES · 4
M  src/lib/auth.ts             +18  -9
A  src/lib/auth.test.ts        +44  -0
M  src/components/Sidebar.tsx   +6  -2
M  README.md                    +7  -1
```

Nothing observed causes a refresh — editing, staging, committing, or switching between sessions/projects.

## Expected

The panel tracks `git status` for the focused session’s working directory and refreshes when the working tree changes — on file save, staging, commit and branch switch — or at minimum offers a manual refresh control.

## Notes

Happy to grab logs or test a build if that helps narrow it down.

metadan:
Fixes #3.

## Diagnosis

The panel is not going stale — it was never reading git. `GitTab.tsx` rendered `PLACEHOLDER_CHANGES`, a fixed four-entry array left over from the design:

```ts
const PLACEHOLDER_CHANGES = [
  { f: "src/lib/auth.ts", p: 18, m: 9, s: "M" },
  { f: "src/lib/auth.test.ts", p: 44, m: 0, s: "A" },
  { f: "src/components/Sidebar.tsx", p: 6, m: 2, s: "M" },
  { f: "README.md", p: 7, m: 1, s: "M" },
];
```

That is exactly the file list, status letters and counts quoted in #3, including the `· 4`. The existing comment says as much: *"Placeholder from the design: the per-file change detail still needs the backend git diff."*

The Branch/upstream section directly above it **is** real, which is what made the placeholder read as a working panel showing stale data rather than as an unimplemented one. Worth noting for anyone triaging: the Changes **modal** has always shown real data, so the two disagreed.

## Change

The backend already exists and already serves the modal — `git::changed_files()` returns `path`, `status`, `additions`, `deletions` and `binary` per file via the `git_changed_files` command — so this is wiring rather than new plumbing:

- Fetch through `gitChangedFiles(path)` and poll it. A working tree changes from outside the app (an editor saving, `git add`, `git commit`, a branch switch) with no event to subscribe to, so polling is the available mechanism; the 3s cadence matches the info tab's process stats. Happy to switch this to a filesystem watcher if you would prefer one — it seemed a larger change than this fix warranted.
- Move the status-to-badge mapping into `components/diff/changeStatus.ts`, shared by the modal and the tab so they cannot drift. Importing it from `ChangesModal` directly would have pulled CodeMirror into the right panel's bundle.
- Add `.gb-R` to the badge colors — it was missing, so renamed files rendered unstyled.
- Show an em dash rather than `+0 −0` for binary files, whose `numstat` carries no line counts.
- Empty and non-repository directories both render `changes.noChanges` instead of a stale list.

Strings go through i18n; `changes.title` and `changes.noChanges` already existed.

## Tests

Six tests in `GitTab.test.tsx`, mocking the IPC layer:

| Test | Covers |
| --- | --- |
| renders backend files and counts | The data is the backend's, not a constant |
| picks up changes after first render | The reported symptom in #3 |
| empty state for a clean tree | No stale list when nothing is changed |
| omits line counts for binary files | `numstat` has no counts to show |
| stays empty outside a repository | The command rejects; no error row |
| no query without a path | No session, no call |

Full suite passes (176 tests, 25 files), `tsc --noEmit` clean, `eslint` clean.

### DIFF
--- a/src/components/diff/ChangesModal.tsx
+++ b/src/components/diff/ChangesModal.tsx
@@ -17,15 +17,7 @@ import {
 } from "../../ipc/commands";
 import { useTermStore } from "../../store/termStore";
 import { languageExtensionFor, vlxCmHighlighting } from "./codeMirrorTheme";
-
-/** Map status to a badge letter and color. */
-const STATUS_META: Record<string, { letter: string; color: string }> = {
-  added: { letter: "A", color: "var(--green, #3fb950)" },
-  modified: { letter: "M", color: "var(--yellow, #d29922)" },
-  deleted: { letter: "D", color: "var(--danger, #e05252)" },
-  untracked: { letter: "U", color: "var(--text-muted)" },
-  renamed: { letter: "R", color: "var(--cyan, #4aa)" },
-};
+import { statusMeta } from "./changeStatus";
 
 /** Render a single-file diff by loading both texts into a read-only MergeView, adding language support transparently once loaded. */
 function DiffView({ cwd, path }: { cwd: string; path: string }) {
@@ -212,7 +204,7 @@ export function ChangesModal() {
               </div>
             )}
             {files?.map((f) => {
-              const meta = STATUS_META[f.status] ?? STATUS_META.modified;
+              const meta = statusMeta(f.status);
               const active = f.path === selected;
               // Split into directory and filename. The directory may truncate with an ellipsis, while the filename
               // remains complete to preserve the most informative tail. Backend paths always use forward slashes
--- a/src/components/diff/changeStatus.ts
+++ b/src/components/diff/changeStatus.ts
@@ -1,0 +1,16 @@
+//! Badge presentation for a `ChangedFile` status, shared by the Changes modal and the right panel's
+//! Git tab so both render the same letter for the same status.
+
+/** Map status to a badge letter and color. */
+export const STATUS_META: Record<string, { letter: string; color: string }> = {
+  added: { letter: "A", color: "var(--green, #3fb950)" },
+  modified: { letter: "M", color: "var(--yellow, #d29922)" },
+  deleted: { letter: "D", color: "var(--danger, #e05252)" },
+  untracked: { letter: "U", color: "var(--text-muted)" },
+  renamed: { letter: "R", color: "var(--cyan, #4aa)" },
+};
+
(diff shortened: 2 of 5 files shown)


vlinx-io:
Thanks for this PR — the diagnosis was exactly right: the panel was rendering `PLACEHOLDER_CHANGES` from the design phase and never reading git. Apologies that it got resolved on a parallel path: a fix for the same issue was already in flight in our internal tree and landed on `main` with the latest sync. The shipped version works much like yours — it polls `git_changed_files` while the tab is visible, refreshes when the window regains focus, adds a manual refresh control, and shares the status-badge mapping with the modal via `changeStatus.ts`.

That parallel fix is why this PR now shows conflicts, so I'm closing it — but one part is still very much worth having: the `GitTab.test.tsx` suite. Our tree currently has no tests for this tab, and your six cases (backend data, pickup after first render, clean-tree empty state, binary files, non-repo, no-path) map cleanly onto the shipped implementation. If you'd like to resubmit just the test suite adapted to current `main`, I'd be glad to merge it under your authorship. Thanks again for the careful report and diagnosis in #3.

PR #27, opened 2026-08-14
feat(web): harden the remote-access management plane

### LINKED ISSUE #24: Harden the remote-access management plane (settings ACL, command allowlist, persistence coordination, login rate limit)
Follow-up to #15/#17 (PRs #21, #23): the reviews there surfaced a set of confirmed hardening gaps in the remote-access management plane. None of them escalate privileges in the current threat model — a paired device already has a full shell via `pty_spawn` — but they weaken the security story of a now-persistent share, and several are cheap to close.

## Primary scope (the four majors)

1. **Key filtering / namespace ACL for the settings store over WS.** `get_app_settings` returns the complete settings map and `set_app_settings` upserts arbitrary keys for every authenticated WS client (`web/dispatch.rs`). Since #23 this includes `remoteAccess.passwordHash` (Argon2id PHC — an offline-bruteforce target) and the autostart keys (`enabled`/`port`), so a paired client can read the verifier or rewrite the password/port that the next restart uses. Filter security-relevant namespaces out of the WS-reachable settings arms (or move them out of `app_settings`).
2. **Source-based allowlist for management-plane commands in `dispatch()`.** Since #21, `web_pairing_create` / `web_devices_list` / `web_device_revoke` (and previously `web_server_stop`) are reachable by every token-authenticated WS client — a paired remote browser can read/rotate the pairing token (rotation also clears the blocklist) or revoke other devices. `dispatch()` should gate management commands by connection source (local shell / Electron loopback vs. remote paired device).
3. **Coordinate multi-writer persistence of `vlx-web-access.json`.** In `serve_main`, the CLI primary server and the auto-started secondary `remote_web` instance each hold their own `AuthState` and write-through the full in-memory state (last-writer-wins): a rotation or revocation can be silently re-persisted away by the sister instance. Share one store handle, or add file locking / re-read-before-save.
4. **Rate-limit + offload unauthenticated password verification.** Argon2id verification (deliberately expensive) runs synchronously on the async executor and is reachable unauthenticated via `POST /api/login` and the WS handshake on `0.0.0.0` — a DoS-amplification primitive. Add a rate limit/lockout and run the verify in `spawn_blocking`.

## Secondary checklist (minors from the same reviews)

- [ ] Create the secret file with `0600` at open time (`OpenOptions.mode`) instead of chmod-after-write (`web/access_store.rs`; same class exists in `e2ee.rs` on baseline — fix both).
- [ ] `--serve` prints the pairing URL including the now-long-lived `#pair` token fragment to stdout on every start (persists in journald across boots). Print without the fragment or gate behind a flag.
- [ ] Cap the persisted device registry (max devices, name length) — currently unbounded, attacker-influenceable growth with write amplification on every registration.
- [ ] Surface `persist()` failures of `block_device` to the caller (currently success is reported even when the revocation only lives until the next restart).
- [ ] chmod `0600` the SQLite DB (it now contains the PHC) for consistency with the token file.
- [ ] Settings-read errors during autostart are logged but never surfaced as `autostartError` in the panel; `autostart_error` also survives a manual stop. Align error surfacing.
- [ ] Startup race: auto-start thread vs. very early manual start can stop each other's instance — skip auto-start when an instance is already running.
- [ ] Doc drift: `Cargo.toml` still justifies `sha2` with the removed salted-SHA-256 login hashing; `DeviceEntry` doc still says "kept in memory for display only"; the comment above `DIRECT_DESKTOP_CMDS` in `transport.ts` predates the dispatch arms.
- [ ] Lockfile hygiene: `argon2` now exists twice in the tree (0.5.3 direct, 0.6.0-rc transitive via russh) — align when practical.

## Non-goals

- No change to the threat model itself (paired device = trusted with shell access stays as-is).
- No token auto-rotation scheme (can be its own follow-up if wanted).

Depends on #21 and #23 being merged (and touches code added by both).

## Acceptance criteria

A paired remote client can no longer read or write the `remoteAccess.*` settings keys nor invoke pairing-management commands; rotation/revocation survive restarts in the dual-instance `--serve` setup; repeated failed logins are rate-limited without blocking the executor; the checklist items are addressed or explicitly deferred with a note.

### LINKED ISSUE #28: Hardening follow-up: secret-ACL gaps, per-instance rate limiter, indistinct rate-limit errors, data_dir reachable via file arms
Review findings surfaced while implementing #24 (retroactive tracking issue; fixed by commit `d82f7e6` on the #27 branch).

## Findings

1. **`gitea_set_config` bypassed the new write-ACL**: it remained remote-dispatchable and wrote the protected `gitea.token` key (keyring, or the plaintext `app_settings` fallback). The secret-command sweep was incomplete as a class: the keyring-deleting `ssh_host_forget`/`url_host_forget` arms were also neither gated nor documented as reviewed exceptions.
2. **The login rate limiter was per server instance** — the dual-instance `--serve` setup granted 2×5 attempts/60 s per IP; `allow`/`record_failure` were not atomic under concurrency.
3. **`wsClient.ts` mapped every WS handshake error to "wrong password"**, discarding the `rate_limited` reason the backend sends precisely so clients can distinguish throttling from a bad password.
4. **The new gating/ACL error strings reached the UI as raw English**, violating the i18n convention.
5. **The generic file arms (`read_text_file`/`write_text_file`/`read_file_base64` and their move/delete/export siblings) let a paired remote client read, overwrite, move or delete the data_dir secret files** (SQLite DB with the PHC, `vlx-web-access.json`, E2EE/TLS keys) — including two-step exfiltration via `rename_path`.

## Fix (shipped)

Secret-command sweep completed (`SECRET_CMDS` gate before the match, stable error codes); recursive fail-closed symlink resolution for the data_dir path ACL; a chokepoint test that parses the production dispatch arms and fails on any path-taking arm without a guard or documented exception; process-wide shared login limiter (weak-registry pattern) with attempt reservation and window pruning; `rate_limited` distinguished in the client; error codes mapped to i18n keys in all 11 locales.

Fixed by `d82f7e6` in #27.

### LINKED ISSUE #29: Share panel: displayed URL/QR ignore live IP selection; CGNAT ordering heuristic; missing stopped-state coverage; 429 latched
QA findings from the #16 implementation, documented as known limitations in #25 (retroactive tracking issue; fixed by commit `7d150b8` on the #27 branch).

## Findings

1. **Interface appearing after server start** (the exact Tailscale scenario from #16): the displayed URL list, copy button and QR were built from the `lan_ips` snapshot frozen at server start, while the selector enumerates live. Selecting an IP whose interface came up after the server started regenerated the pairing link correctly, but the primary displayed URL/QR fell back to the snapshot until a server restart.
2. **"CGNAT behind LAN" ordering** was only guaranteed via the platform-dependent broadcast-capability heuristic, never by `is_cgnat` itself, and the invariant had no deterministic test.
3. **The stopped-state IP selector had no test assertion**, and two timing races existed: the link-regeneration effect only fired when `pairUrl` was already set (a persisted IP arriving during the in-flight initial pairing never re-paired), and `genPairing` had no in-flight guard (overlapping calls could let a stale response win).
4. **HTTP 429 during silent relogin was latched like a password rejection**, permanently stopping retries although the rate-limit window expires after 60 s.

## Fix (shipped)

Share URLs derive from the live selection (scheme from the mode, port from status) independent of the start snapshot; explicit CGNAT-behind-everything sort key, unit-tested as a pure function; stopped-state coverage plus a sequence guard and a corrected regeneration effect; 429 no longer latches (shows `login.rateLimited`, later attempts can succeed).

Fixed by `7d150b8` in #27.

### LINKED ISSUE #30: Residual robustness sweep: path-ACL chokepoint gaps, limiter cancellation leak, crash-unsafe persistence
Residual robustness findings from the reviews of the #24 hardening work (retroactive tracking issue; fixed by commit `312be1d` on the #27 branch).

## Findings

1. **Path-ACL chokepoint gaps**: the completeness test's `PATH_KEYS` vocabulary missed `rootPath` and `folderName`, and the arm parser did not recognize multi-pattern arms — the test asserted more than it guaranteed.
2. **Check/use divergences**: `save_doc_image` gated the proxy path (`docPath`) instead of the effective `<parent>/assets/<name>` write location; `git_file_diff` computed its effective path twice (gate and read could diverge, plus a redundant git subprocess on the local lane).
3. **Rate limiter**: the pending reservation leaked for up to 60 s when a request future was cancelled mid-verify; `record_success` dropped the whole per-IP window entry, refreshing an attacker's failure budget behind shared NAT.
4. **Persistence crash-safety**: `access_store::save` did not fsync the tmp file or directory around the rename; the process-wide `PairingState` registry silently fell back to the raw path when `canonicalize` failed (two instances could get different keys for the same directory); `rotate`/`block` mutated memory before persisting (memory/disk divergence on persist errors).
5. **Smaller items**: a malformed base64 handshake frame threw uncaught in `wsClient.ts` instead of failing the handshake; the restart test proved less than its name claimed since the shared registry landed; incomplete pairing-state reset paths; the synthesized share-URL scheme was inferred from a snapshot URL instead of being reported by the backend; two docstring/test drifts; the headless test suite was only compiled, never executed, during verification.

## Fix (shipped)

All of the above addressed, including a parser negative-test, RAII drop-guard release with surgical success accounting, platform-correct fsync (a first attempt via a read-only handle would have broken Windows — caught by the review round), fail-closed registry keying, persist-before-commit, an explicit `scheme` field in `WebServerStatus` (both transports), and executed headless suites in verification. Inherent limits (hard links / check-then-use window of the prefix ACL; TLS SAN gap for live-selected IPs) are documented in code as accepted — tracked separately.

Fixed by `312be1d` in #27.

mguttmann:
Fixes #24. Fixes #28. Fixes #29. Fixes #30.

> **Merge order:** this branch merges the other three PRs (#21 → #23 / #25) and builds on top. Merge those first; the diff then shrinks to the hardening changes plus the (session-resolved, fully test-verified) merge of the three branches.

## Changes (the four majors from #24)

1. **Settings ACL over WS**: `remoteAccess.*` and `gitea.token` are filtered from remote `get_app_settings` reads; a remote `set_app_settings` batch containing a protected key errors without touching the DB. Local callers are unaffected.
2. **Source-based command gating**: `dispatch()` gains a `CallOrigin::{Local,Remote}` parameter — Local is the Tauri `desktop_call` path and clients of a `LoopbackHttp` (Electron sidecar) instance; Remote is LanTls/LanHttp. The whole management set (`web_server_start/stop/status`, `web_pairing_create`, `web_devices_list`, `web_device_revoke`, `network_interfaces_list`) is rejected for remote clients before any argument parsing. Regression tests cover both origins.
3. **Coordinated persistence**: all in-process server instances share one `PairingState` per `data_dir` (process-wide weak registry), eliminating the last-writer-wins clobber between the CLI primary and the auto-started secondary instance in `--serve`. Restart persistence remains covered by tests that drop the first instance.
4. **Login rate limit + executor offload**: per-IP fixed window (5 failures / 60 s) checked **before** any Argon2 work — `/api/login` answers 429, the WS handshake answers `rate_limited`; Argon2id verification runs in `spawn_blocking` behind a process-wide `Semaphore(2)`, and without a valid pairing token no Argon2 runs at all (the sync verify only exists under `#[cfg(test)]`, so the invariant is compiler-enforced).

**Checklist from #24:** (a) secrets now created `0600`-at-open (`access_store` **and** the pre-existing `e2ee.rs` instance); (b) `--serve` prints the pairing URL tokenless unless `--print-pairing` is passed or stdout is a TTY; (c) device-registry caps (32 devices / 64+128 char fields); (d) persist failures surfaced as `Result` for `block_device` and `rotate_pairing_token`; (e) SQLite DB chmod 0600; (f) autostart settings-read errors surface as `autostartError`, and a manual stop clears a stale error; (g) autostart skips when an instance is already running; (h) the three doc-drift comments fixed. **Deferred with rationale:** (i) argon2 version dedup (0.6 is still rc, transitive via russh) and a blocklist cap (after gating, only local callers can grow it).

## Review findings surfaced (for maintainer awareness; none block per the threat model — a paired device retains full shell via `pty_spawn`)

- **[major] `gitea_set_config` bypasses the new write-ACL**: it remains remote-dispatchable and writes `gitea.token` (keyring, or plaintext `app_settings` fallback) — the very key the ACL protects on the settings arm. One-line follow-up: add it to the remote-gated set (or gate on origin inside the arm).
- The **rate limiter is per server instance** (unlike the deliberately shared `PairingState`), so the dual-instance `--serve` setup offers 2×5 attempts/60 s per IP; keying is by bare source IP (IPv6 rotation widens the budget; shared NAT can lock out neighbors — the window is short).
- `allow → verify → record_failure` is not atomic: N parallel requests from one IP can all pass the check before the first failure is recorded (soft, not hard, limit under concurrency).
- **Client contract drift**: the WS handshake now sends `rate_limited` so clients could distinguish throttling from a wrong password, but `wsClient.ts` still maps every handshake error to "wrong password" UX; the new backend gating/ACL error strings are also raw English (i18n follow-up).
- **Accepted residual risk (named, per non-goal):** the generic file arms (`read_text_file`/`write_text_file`/`read_file_base64`) still let a paired remote client read/write the SQLite DB and token file directly — closing that means sandboxing the file API, a different ticket.
- `rotate`/`block` mutate memory before persisting; on a persist error the live state and disk diverge until restart (error is now at least surfaced).

## Verification

- `cargo test` **388 passed** (new: ACL filter/reject, origin gating both ways, shared-store rotation/revocation across instances, 429-before-Argon2 integration test, cap/eviction, 0600 modes — all loopback-only), `pnpm test`, `pnpm build`, headless `cargo build --no-default-features` — green. `pnpm lint` fails only on the pre-existing baseline errors in `src/remote/ConnectionBanner.tsx`.
- The WS-handshake rate-limit branch is review-verified but has no automated test (documented infra limit: the existing test harness drives the HTTP login path).

Closes the series: #21 (dispatch), #23 (persistence), #25 (IP selection), this (hardening).

## Follow-up commits on this branch (post-review)

- `d82f7e6` — fixes #28: secret-ACL gaps closed as a class, process-wide shared login limiter, `rate_limited` distinguished in the client, error codes mapped to i18n, data_dir shielded from the generic file arms (chokepoint test).
- `7d150b8` — fixes #29: share URLs follow the live IP selection, deterministic CGNAT ordering, stopped-state coverage and race guards, 429 no longer latches silent relogin.
- `312be1d` — fixes #30: chokepoint parser hardened, effective-path gating, RAII limiter reservations, crash-safe persistence, handshake robustness, explicit `scheme` in the status.

Deliberately documented rather than fixed (see the summary comment): tracked in #31.


mguttmann:
Three follow-up commits landed on this branch after review:

- **d82f7e6** — ACL gaps closed (secret-command sweep incl. `gitea_set_config` and the keyring-forget arms; recursive fail-closed symlink resolution; a chokepoint test that parses the production dispatch arms and fails on any ungated path-taking arm), process-wide shared login limiter with attempt reservation, `rate_limited` distinguished in the client, stable error codes mapped to i18n keys (11 locales).
- **7d150b8** — share URLs derive from the live IP selection (an IP that appears after server start now drives copy URL, pairing link and QR), deterministic CGNAT-behind-LAN ordering as a pure tested function, stopped-state selector coverage plus race guards, 429 no longer latches silent relogin.
- **312be1d** — residual sweep: chokepoint parser hardened (multi-pattern arms, missing path keys), effective-path gating for `save_doc_image`/`git_file_diff`, RAII release for limiter reservations with surgical success accounting, crash-safe persistence (fsync around rename, fail-closed registry keying, persist-before-commit), handshake robustness on malformed base64, stricter restart test, explicit `scheme` in `WebServerStatus`; inherent limits (hard links / check-then-use of the prefix ACL, TLS SAN gap for live-selected IPs) are documented in code as accepted.

Known remaining trade-offs, documented rather than fixed (diminishing returns; maintainer call): the chokepoint test is a textual analysis and guarantees less than a typed dispatch table would (structural rework candidate); cancelling a login mid-verify frees the rate-limit booking while the non-cancellable Argon2 task finishes under the process-wide `Semaphore(2)` — actual Argon2 concurrency stays bounded at 2, only the bookkeeping is optimistic; the Windows fsync path is compile-checked but has no executing Windows test lane here.

PR #51, opened 2026-08-23
fix: keep color-scheme in sync with the selected theme

FarhadGSRX:
## Problem

`applyTheme()` writes `data-theme` onto `documentElement`, but never updates `color-scheme`. `index.html` seeds that property once, from the OS:

```html
:root { color-scheme: dark; }
@media (prefers-color-scheme: light) { :root { color-scheme: light; } }
```

Since nothing updates it afterwards, the seeded value sticks for the window's lifetime. Choosing a theme that disagrees with the OS therefore leaves every native in-page control — checkboxes, radios, selects, scrollbars, form fields — painted in the OS scheme while the rest of the UI switches.

Easiest repro, on a dark-mode OS: pick **Light** in the titlebar theme selector, then quit. The "save workspace" checkbox in the quit dialog is still dark, on an otherwise light dialog.

## Fix

Set `documentElement.style.colorScheme` alongside `data-theme`. The inline style outranks the `:root` rule in `index.html`, and `system` mode keeps tracking the OS because `watchSystemTheme` → `applyAppearance` already re-runs `applyTheme` on change.

## Verification

- Confirmed by hand on Windows (WebView2), in both directions: dark → light and light → dark, the native checkbox follows the app.
- Adds `src/theme.test.ts` covering both explicit modes and `system`. Reverting the one-line change makes all three tests fail, so they measure the defect rather than the code.
- `tsc --noEmit` is clean.
- `eslint .` is clean (exit 0, no output) and `tsc --noEmit` is clean. Both were measured after rebasing onto v0.1.102, which fixed the two `ConnectionBanner.tsx` errors that `main` used to carry — so the lint bar is zero now and this meets it, rather than merely matching a red baseline as an earlier revision of this text said.
- The remaining `vitest` failures on my Windows box are 5s timeouts rather than assertion failures, and reproduce without this branch; the count varies run to run, so they are load-dependent on that machine, not caused here.

### DIFF
--- a/src/theme.test.ts
+++ b/src/theme.test.ts
@@ -1,0 +1,42 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// applyTheme pushes the resolved palette into the live terminal registry; stub it so the module can be
+// exercised without any xterm instances.
+vi.mock("./terminal/registry", () => ({ setXtermTheme: vi.fn() }));
+
+import { applyTheme } from "./theme";
+
+/** jsdom has no matchMedia; report the OS as the given scheme so 'system' resolves predictably. */
+function mockSystemScheme(scheme: "dark" | "light") {
+  vi.stubGlobal("matchMedia", (query: string) => ({
+    matches: query.includes("dark") === (scheme === "dark"),
+    addEventListener: vi.fn(),
+    removeEventListener: vi.fn(),
+  }));
+}
+
+describe("applyTheme", () => {
+  beforeEach(() => {
+    document.documentElement.removeAttribute("style");
+    document.documentElement.removeAttribute("data-theme");
+    localStorage.clear();
+  });
+
+  // color-scheme decides how the user agent paints native in-page controls. index.html seeds it from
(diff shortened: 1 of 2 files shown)


FarhadGSRX:
Rebased onto v0.1.102.

Worth noting the interaction with the `syncNativeChrome()` / `set_native_theme` work that landed in that release, since both touch `applyTheme` and it would be easy to read them as the same fix. They are different layers:

- `syncNativeChrome` tints the **native window chrome** — the OS title bar, via DWM on Windows.
- This PR sets CSS `color-scheme`, which is what the WebView uses to paint **native in-page controls**: checkboxes, radios, selects, scrollbars, form fields.

The second is still unset after the release, so forcing a theme against the OS scheme still leaves those controls in the OS scheme. Easiest repro on a dark-mode OS: pick Light in the titlebar selector, then quit — the "save workspace" checkbox in the quit dialog is still dark on an otherwise light dialog.

The conflict was positional only; both calls now sit in `applyTheme` and are independent.

PR #64, opened 2026-09-10
feat(chat): open background tasks as live tabs in the conversation view (#63)

### LINKED ISSUE #63: Conversation view: background tasks (workflows, subagents) cannot be opened or inspected although the protocol delivers the full progress tree
## Summary

In the experimental conversation view, a running background task (a Claude Code workflow with dozens of agents, a subagent, a backgrounded shell command) is only visible as one truncated line in the "Tasks" chip at the bottom right, with a Stop button. Nothing can be opened. There is no way to see which phase runs, which agents are active or done, how long they have been running or how much they consumed. A workflow that legitimately runs for one or two hours becomes a black box; the user can only trust that it finishes.

In the terminal view the same run is fully inspectable through Claude Code's own `/tasks` and `/workflows` views, so users who rely on the conversation view lose something they had before.

Switching back to the terminal view is not an alternative while the run is in progress: the conversation view drives Claude as a stream-json peer, so the switch stops the running turn (the confirmation dialog says so). That part is by design and not the subject of this issue.

## What the protocol already delivers

Verified on Claude Code 2.1.267 with a three-agent probe workflow in `--output-format stream-json`. Every `task_progress` frame of a `local_workflow` task carries the complete state that the TUI's `/workflows` view renders:

- `description` is live and has the form `<phase title>: <agent label>`, `last_tool_name` is the current agent label, `usage` has `total_tokens`, `tool_uses`, `duration_ms`.
- Roughly every second frame carries `workflow_progress`, an array of `{ type: "workflow_phase", index, title }` and `{ type: "workflow_agent", index, label, phaseIndex, phaseTitle, agentId, model, state ("start" | "done"), startedAt, queuedAt, attempt, promptPreview, lastProgressAt, tokens, toolCalls, durationMs, resultPreview }`.
- `task_started` carries `task_type: "local_workflow"`, `workflow_name`, `description` and the script as `prompt`; `task_updated` and `task_notification` close the task with status, `end_time`, `summary` and final `usage`.

Excerpt of one real frame (trimmed):

```json
{"type":"system","subtype":"task_progress","task_id":"worwyzf75","tool_use_id":"toolu_01Tpo…",
 "description":"Beta: gamma-worker","last_tool_name":"gamma-worker","summary":"protocol probe",
 "usage":{"total_tokens":64131,"tool_uses":0,"duration_ms":6326},
 "workflow_progress":[
   {"type":"workflow_phase","index":1,"title":"Alpha"},
   {"type":"workflow_phase","index":2,"title":"Beta"},
   {"type":"workflow_agent","index":1,"label":"alpha-worker","phaseIndex":1,"phaseTitle":"Alpha",
    "model":"claude-opus-5[1m]","state":"done","startedAt":1789036353475,"attempt":1,
    "promptPreview":"Reply with the single word ALPHA…","tokens":21379,"toolCalls":0,"durationMs":2094,"resultPreview":"ALPHA"},
   {"type":"workflow_agent","index":3,"label":"gamma-worker","phaseIndex":2,"phaseTitle":"Beta","state":"start", "...":"..."}
 ]}
```

Only the workflow's `log()` narrator lines are not part of the protocol.

## Where it is dropped today

- `src-tauri/src/agent/chat/engine.rs`, `handle_claude_task`: after `task_started`, only `usage` and the terminal status are folded into the Workflow card. The live `description`, `last_tool_name` and the whole `workflow_progress` array are ignored.
- The Tasks chip (`src/layout/CenterPane/session/extras.tsx`, `TasksChip`) lists the static inventory from `background_tasks_changed` (`task_id`, `task_type`, `description` as given at start), truncated by CSS ellipsis, with no way to open the task and no link to the card in the timeline.
- `ChatBackgroundTask` in `src/ipc/chat.ts` has no room for progress at all.

## Proposal

1. Make each entry in the Tasks chip openable. Opening a task adds a tab at the top (the same place the Memory library opens as a tab today), so two, three or five parallel tasks can each have their own tab and stay open while the conversation tab is in front.
2. The task tab shows, live: the task description and type, elapsed time, total tokens and tool uses, the current status, and for workflows the phase list with every agent (label, phase, model, state, attempt, duration, tokens, prompt preview and, when done, result preview). Stop stays available from the tab.
3. Keep the live `description` in the chip row itself so the "phase: agent" line is visible without opening anything.
4. Optionally let the Workflow card in the timeline open the same tab.
5. Stay fail-soft: older CLI versions without `workflow_progress` still get the description, usage and status view.

I am happy to build this and open a pull request against `main`; the branch is in progress on my fork.

mguttmann:
Closes #63

## What

In the experimental conversation view, a running background task (a Claude Code workflow with many agents, a subagent, a backgrounded shell command) was one truncated line in the Tasks chip with a Stop button. Nothing could be opened. This PR makes each task openable as a tab of its own and renders the state that Claude Code already streams.

- **Tasks chip:** each row shows the live description (`Phase: agent` for workflows), the task type, and its status once finished. The badge counts running tasks only. Finished tasks stay listed with their final state until the next turn starts, so their result can still be opened.
- **Task tab:** clicking a row opens a tab in the TabBar (parallel to document and browser tabs, client-local, not persisted, not mirrored). Several tasks open as several tabs, each closable on its own; tabs stay mounted when another tab is in front. The tab shows type, status, elapsed time (ticking while running), total tokens, tool uses, the current agent, start and end time, the final summary and output file, and for workflows the phase list with every agent: label, model, state, attempt, duration, tokens, prompt preview and result preview.
- **Lifecycle and mirror mode:** when the Claude process exits or a new one takes the session over, an open task tab shows the task as ended instead of a running clock and a Stop that could only fail. This holds for as long as the tab receives the session's chat events: on the Tauri transport always; on the WebSocket transport until a reconnect happens while no conversation pane for that session is mounted any more (the tab does not re-register the session's events itself yet, see the open points). A task tab that outlives its conversation pane keeps getting the task's final status, summary and output file: the process is no longer released on the inventory drop itself but only after the task's `task_notification`, bounded by a short grace (`TASK_SETTLE_GRACE`) so a task whose frames never come cannot keep the process alive. In mirror mode a client that sits in a task tab publishes a session tab in its place (with that tab's session and first pane), so the peer never sees an empty stage; and the client itself stays in its task tab whatever an incoming frame names as active, on both sides (publisher receiving the reconcile echo or an ordinary peer push, follower receiving the publisher's pushes) until the user leaves the task tab.
- **Rust engine:** `ClaudeExtras.background_tasks` becomes a typed, merged view per `task_id` over `background_tasks_changed`, `task_started`, `task_progress`, `task_updated` and `task_notification` (live `description`, `last_tool_name`, `usage`, `workflow_progress`, `started_at`, `ended_at`, `status`, `summary`, `output_file`). A task the inventory drops before its terminal frame arrives (the real wire order, see below) is marked `ended` instead of vanishing. Finished tasks are pruned when the next user turn starts, capped at 20. No new command, no transport change: the existing `extras` snapshot and event carry it on Tauri IPC and WebSocket alike.

## Why the data is there

Verified on Claude Code 2.1.267 with a three-agent probe workflow in `--output-format stream-json`: every `task_progress` frame of a `local_workflow` task carries the live `description`, `last_tool_name`, `usage` and, about every second frame, `workflow_progress`, the same phase and agent tree the TUI renders under `/workflows`. The Rust tests replay those frames in wire order (inventory empties before `task_updated` and `task_notification` arrive).

## Scope and non-goals

- No version bump, no changelog entry, no change to `protocol.rs`, `command_core.rs`, `commands.rs` or `web/dispatch.rs`.
- Task tabs are not persisted across restart (the process behind them is gone by then) and are not published to mirror peers.
- Workflow `log()` narrator lines are not part of the protocol and are not shown.
- Foreground subagents that never appear in the background inventory are held but not listed, as before.
- The Workflow card in the timeline does not open the tab yet.
- 23 new i18n keys (`chat.tasks.*`) in all 11 locales; German translated, the others carry the repo's `// TODO translate` marker.

## Known minor points left open

- "Finished" is defined twice (Rust allow-list, TS deny-list); an unknown future terminal status would read as finished in the UI and as running in Rust.
- The cap of 20 finished tasks also counts unlisted foreground subagents.
- A task marked `ended` because the inventory dropped it is not revived if the inventory lists it again without a status.
- Elapsed time while running mixes the backend's `started_at` with the client clock; on a remote browser with clock skew the figure can be off until the task ends.
- The phase and agent lists in the task tab use the wire's `index` as React key; the TS types declare it required, the Rust side forwards `workflow_progress` unvalidated, so a frame without it would warn in dev.
- `debug.ts` tab invariants do not know task tabs (pre-existing class: they do not know browser tabs either).
- When a publisher opens a task tab, the peer is moved to the derived anchor's first pane (and, if the publisher came from a document or browser tab, to the last session tab), not to the pane it mirrored before.
- Browser/Electron transport: a task tab whose conversation pane is gone does not re-register the session's chat events after a WebSocket reconnect; until the pane is opened again it stops receiving updates.
- A `task_started` or a `task_updated` back to running for an already finished task id keeps `ended_at`, summary and output file from the previous life of that id (frozen elapsed time next to a Running badge); task ids are random per process, so this is not reachable in practice.

## Tests

- Rust: 7 tests in `engine.rs` replaying the probe frames (merge, inventory drop before terminal frame, detached release waiting for the terminal frames of a dropped task, release after the grace when no frames come, pruning and cap, unlisted foreground subagents, lenient frames). `cargo test`: 837 passed, 0 failed.
- Frontend: new `taskTabs.test.ts`, `TaskView.test.tsx`, `TabBar.taskTab.test.tsx`; extended `extras.test.tsx`, `mirrorLayout.test.ts`, `mirrorSync.test.ts` (reconcile round trip through the real store; a client in a task tab stays there as publisher and as follower). `pnpm test`: 651 passed. `pnpm build` and `pnpm lint` green.
- Not verified end to end: the browser/Electron transport with a live WebSocket, a manual run with a real multi-agent workflow in the desktop app, the look of the chip row after the split into two sibling buttons, and the real width of the flush window between Claude's inventory drop and its terminal frames (the grace length is a setting, not a measurement).


mguttmann:
Rebased onto v0.2.0 (`af9e7de`), no functional change. Conflict resolution: v0.2.0 brought its own tasks chip into `ChatPane.tsx`, so the branch now only wires the `onOpen` handler into that chip instead of adding a second one; the task view CSS is appended after the v0.2.0 block, and the TabBar test mock follows the `dateLocale` to `getLocale` rename. On the rebased branch: `cargo test` 1038 passed, `pnpm test` 116 files / 1039 tests passed.

mguttmann:
Rebased onto v0.2.2 (7874081). One conflict, the chat import line in termStore.ts (v0.2.2 added chatSend and chatStart for auto-continue, this branch adds the ChatBackgroundTask type); both kept. pnpm test 120 files / 1070 tests, build and lint green on the rebased branch.

PR #70, opened 2026-09-12
fix(chat): always state the Claude process's permission mode (#69)

### LINKED ISSUE #69: Conversation view: the chat process inherits the host's permission defaultMode, and the view's Bypass switch is rejected at runtime
## Summary

In the experimental conversation view, the Claude process the chat engine spawns does not carry the permission mode the view displays, and the view's own mode switch cannot move it into bypass. Both are one class: the engine leaves the permission mode to the CLI's defaults instead of stating it, and the CLI's defaults are made for a TUI, not for a protocol peer whose permission prompt is VelaTerm itself.

Observed on `main` at `cf62db9` with Claude Code 2.1.269 on macOS.

## Defect 1: the process inherits the host's `permissions.defaultMode`, the view shows "Default"

`protocol::launch_args` (`src-tauri/src/agent/chat/protocol.rs:76-79`) adds `--permission-mode` only when the session has a stored mode. A session that never had one set in VelaTerm (for example one that was always launched from the terminal view, or with the user's own shell alias) starts the chat process with no mode flag at all. Claude Code then falls back to the user's `~/.claude/settings.json`. With `"permissions": {"defaultMode": "dontAsk"}` there, which is a common setting for people who run headless automation, every Bash and WebFetch call is refused without a question:

```
Permission to use Bash has been denied because Claude Code is running in don't ask mode.
```

The `--permission-prompt-tool stdio` that the engine passes never gets a request, so VelaTerm's permission card never appears. Meanwhile `initial_mode` (`src-tauri/src/agent/chat/engine.rs:960`) reports `default` for an unset mode, so the view shows "Default" while the process runs in `dontAsk`. The user sees a Claude that refuses to work in a view that claims it would ask.

Reproduced directly against the CLI, in a temp directory, with the same flags the engine uses (`-p --verbose --input-format stream-json --output-format stream-json --permission-prompt-tool stdio`) and a prompt that runs `curl` through Bash:

| Launch | `init.permissionMode` | `can_use_tool` requests reaching stdin | Bash ran |
|---|---|---|---|
| no `--permission-mode` (engine today, unset session mode) | `dontAsk` | 0 | no, denied as above |
| `--permission-mode default` | `default` | 1 | yes, after the peer allowed it |

## Defect 2: the view offers "Bypass permissions", the CLI rejects the switch

The mode menu in the chat view sends `set_permission_mode` as a control request (`ChatManager::set_mode`, `engine.rs:1590-1596`). For `bypassPermissions` the CLI answers:

```
Cannot set permission mode to bypassPermissions because the session was not launched with --dangerously-skip-permissions
```

The engine surfaces that as an error and the mode stays. So the one control the view has for this cannot get the user out of defect 1. Restarting into bypass would work in principle (the choice is persisted as `skip`, and a fresh launch maps it to `--permission-mode bypassPermissions`), but nothing in the view does that restart, and the terminal side is unaffected either way.

The CLI has a flag for exactly this situation: `--allow-dangerously-skip-permissions` ("Enable bypassing all permission checks") permits the later switch without starting the process in bypass. Reproduced: launched with `--permission-mode default --allow-dangerously-skip-permissions`, then `set_permission_mode bypassPermissions` over stdin: the control response is `success`, `init.permissionMode` reports `bypassPermissions`, and Bash runs without a request.

## Proposed fix

Both defects close at one chokepoint, `launch_args` for Claude:

1. Always pass `--permission-mode <mode>`, where `<mode>` is exactly what `initial_mode` reports to the view (`default` when the session has none). A protocol peer must never leave its permission mode implicit: the host's `defaultMode` is a TUI setting, and `dontAsk` in particular exists for unattended runs, which is the opposite of a view whose whole point is to ask.
2. Always pass `--allow-dangerously-skip-permissions`, so the view's "Bypass permissions" choice can be applied live through the control request the view already sends. This does not start anything in bypass; it only lets the switch the user makes in VelaTerm's own UI take effect.

Terminal sessions are not affected: they launch through `inject.rs` with `--dangerously-skip-permissions` only when the session's mode is `skip`, as before.

I have a PR ready for this and will link it here.

mguttmann:
Closes #69

## What

The chat engine now states the Claude process's permission mode on every launch, with the same word the view shows. Three files under `src-tauri/src/agent/chat/`, no UI change, no new command, no transport change.

- `protocol::launch_args` takes the permission mode as a required value and always passes `--permission-mode <mode>`. The engine passes exactly the word `initial_mode` reports to the view (`default` when the session has none, `bypassPermissions` for a stored `skip`, `plan`/`acceptEdits`/`auto` through), from one local variable used for both, so the mode on screen and the mode in the process cannot drift apart. The `--dangerously-skip-permissions` flag for a stored bypass mode stays as v0.2.0 added it.
- The skill catalog probe in `skills.rs`, the second caller of `launch_args`, states `default` for the same reason: a launch never leaves the mode to the CLI's settings.

## Why

See #69 for the evidence chain. In short, against Claude Code 2.1.269 with the engine's own flags: with no `--permission-mode`, a user whose `~/.claude/settings.json` says `defaultMode: dontAsk` gets `init.permissionMode = dontAsk`, zero `can_use_tool` requests reach the stdio peer, and Bash is refused with "running in don't ask mode" while the view shows "Default". With `--permission-mode default`, one request reaches the peer and the command runs after it is allowed.

## Rescoped after v0.2.0

The first version of this PR also passed `--allow-dangerously-skip-permissions` so that the view's switch to bypass would be accepted at runtime. v0.2.0 solves that part differently: `permission_restart` recognises the CLI's refusal and asks the user to restart the process in bypass mode. This PR therefore drops that flag and keeps only the first defect, which v0.2.0 still has: `engine.rs` computes `initial_mode(...)` with a `default` fallback for the view but launched the process with `cli_permission_mode(...)`, which is `None` for an unset mode, so the flag was omitted and the process inherited the TUI default.

One vocabulary note: `default` is no longer listed among `--permission-mode`'s documented choices in 2.1.269 (`manual` is the listed spelling) but is accepted as an alias and echoed back as `default` in the init frame. The engine already used that word for `set_permission_mode` before this change; if the alias is ever dropped upstream, `cli_permission_mode` is the single place to remap it.

## Tests

- `protocol.rs`: `launch_args_always_ask_for_streaming_and_permission_routing` asserts that `--permission-mode default` is present and `--dangerously-skip-permissions` is not; the two other `launch_args` tests are adapted to the required parameter.
- `engine.rs`: new `claude_launches_with_the_mode_the_view_reports` walks every stored value (`None`, empty, unknown, `skip`, `plan`) and checks that `initial_mode` and the launch flag agree.
- `cargo test` on the rebased branch: 1032 passed, 0 failed, 15 ignored.
- Not verified end to end in this PR: a full chat turn in the desktop app after the change. The argument vectors match the ones in the reproduction exactly.

## Linking

For maintainers: this PR comes from a fork without write access here, so GitHub does not add it to the issue's Development sidebar on its own even though the body says `Closes #69`. Connecting it from the PR sidebar takes one click; merging closes #69 either way.


mguttmann:
Rebased onto v0.2.0 and rescoped to the first defect only. v0.2.0 handles the runtime switch to bypass through `permission_restart` (restart confirmation), so the `--allow-dangerously-skip-permissions` flag is gone from this PR. What remains: `launch_args` always states `--permission-mode`, with the same word `initial_mode` already reports to the view. The PR text is updated accordingly; `cargo test` on the rebased branch: 1032 passed, 0 failed.

mguttmann:
Rebased onto v0.2.1 (68a4bc2). One conflict: the launch now goes through crate::host::command (your Windows console fix), kept as is with the explicit mode passed into launch_args. v0.2.1 still launches Claude without --permission-mode when nothing is stored (engine.rs passes cli_permission_mode(...) straight through), so the defect in #69 is still present there. cargo test on the rebased branch: 1045 passed.

PR #76, opened 2026-09-13
feat(chat): let users choose the inline composer chips and show More only on real overflow (#71)

### LINKED ISSUE #71: Conversation view: let users choose which composer chips stay inline, and show "More" only when they really do not fit
## What happens today

In the conversation view the composer toolbar shows a fixed set of chips inline (model, effort, permission mode) and hides the rest (fast mode, service tier, personality, MCP, tasks, account) behind a "More" button. The split is hardcoded in `ChatPane.tsx` (the `primary` and `secondary` props around lines 2050 to 2153), and `ComposerToolbar.tsx` renders the "More" toggle on desktop unconditionally (line 45: `{!mobile && <button ... sv-more-toggle>`), whether or not the row has room. Nothing measures the available width.

The result on a normal desktop window: the four chips end well before the middle of the row, the rest of the line is empty, and the chips a user actually reaches for (tasks, MCP, account) still take two taps. The "More" button occupies the space one of those chips could have taken.

## What I would like

1. **A setting that decides which chips stay inline.** Under Settings, Conversation view: a list of the available composer chips with a toggle each (and ideally an order). Chips a user turns on are rendered inline; chips a user turns off go into "More".
2. **"More" only on real overflow.** The toolbar measures the width it has and lays out the enabled chips in order; the button appears only when the enabled chips do not fit, and then holds exactly the ones that did not fit (plus the ones the user turned off). When everything fits, there is no "More" button at all.
3. **Sensible defaults** that match today's inline set, so nothing changes for people who do not touch the setting.

Chip availability already depends on agent kind and engine state (for example MCP only for Claude and Codex while the engine runs, fast mode only when offered), so the setting should be a preference over the chips that exist for the current session, not a promise that a chip is always shown.

## Notes for an implementation

- `ComposerToolbar` is the natural chokepoint: give it the ordered list of enabled chips and let it measure (ResizeObserver on `.sv-controls-primary`, chip widths from the rendered nodes) instead of receiving a pre-split `primary` and `secondary`.
- The setting can live in `src/store/settings.ts` next to the other conversation view preferences and sync like they do.
- Keyboard and screen reader behaviour of the current "More" row (Escape closes, outside click closes, `aria-expanded` and `aria-controls`) should stay as it is when the button does appear.

I am happy to build this and will link a PR here.

mguttmann:
Closes #71

## What changes

- **A setting decides which chips stay inline.** Settings > Conversation view > Composer toolbar lists every composer chip with an On/Off switch and, for chips that are on, up/down arrows for their order. The default is exactly the previous inline set (model, thinking effort, collaboration mode, permission mode), so nothing changes for anyone who does not touch it.
- **More only on real overflow.** `ComposerToolbar` now receives the chips that exist for the session plus the user's ordered list. It lays the enabled chips out in that order and measures them in a layout effect before paint, then follows a `ResizeObserver`. The More toggle is rendered only while the enabled chips do not fit the row; it then holds the chips that overflowed, followed by the chips that are off. When everything fits there is no More button at all.
- **Pure layout logic with its own tests.** `composerLayout.ts` holds the split by preference (`arrangeComposerChips`) and the width partition (`partitionComposerChips`) without React or DOM; the toolbar injects measured widths, the tests inject made-up ones.
- **One chip list instead of a fixed split.** The hardcoded primary/secondary split in `ChatPane.tsx` becomes a single list of chips with ids. Availability per agent kind and engine state stays where it was (MCP and tasks only while the engine runs, fast mode only when offered, account chip yields to the inline sign-in panel, and so on). A chip that is on is always drawn when it exists for the agent kind; momentary unavailability shows as an empty or disabled chip, never as absence (see the rule below).
- **Mobile unchanged in shape.** The mobile layout keeps its two rows (enabled chips first, the rest below) without measuring.
- **Persistence.** `composerInlineChips` joins the settings payload and syncs like the other conversation view preferences. It is sanitized on load: unknown ids are dropped, a malformed value falls back to the default, an explicit empty list is allowed.
- 15 new i18n keys in all 11 locale files.
- The user manual (`docs/manuals`) gains the new Conversation view section and the account chip description follows the new behaviour.

Keyboard and screen reader behaviour of the More row (Escape closes, outside click closes, `aria-expanded` and `aria-controls`) is unchanged when the button does appear.

## Assumptions worth a look

- The setting is global, not per agent kind or per session.
- The Codex reset credits control is treated as a chip like the others (off by default, as before).
- Chips do not shrink: a chip that no longer fits moves into More instead of being squeezed.
- Turning every chip off is allowed; the toolbar then shows only More with everything inside.

## The visibility rule, after a first round of use

Manuel switched the Tasks chip on and saw nothing: the chip hid itself while the task list was empty, and MCP and Tasks were offered only while the Claude process was alive, so they came and went between turns. A chip nobody can find defeats the setting. The branch applies one rule: a chip the user turned on is always visible when it exists for the agent kind. Tasks with an empty list shows its label and "No background tasks"; MCP and Tasks are disabled while no agent process runs (dimmed, not openable, a tooltip names the cause, no status request is sent, no badge); the account chip stays in the row disabled during an unresolved sign-in instead of disappearing. `ChipPopover` gained a `disabled` prop with the existing disabled chip look. Only chips that do not exist for the agent kind stay absent (for example Codex reset credits in a Claude session). The settings hint and the manual describe the rule.

## Verification

On this branch: `pnpm lint` clean, `pnpm build` green, `pnpm test` 120 files / 1053 tests passing on the rebase onto v0.2.2. New tests cover the layout arithmetic, the toolbar's fold behaviour, the settings section, and the sanitizer.

Note on linking: GitHub does not show the Development sidebar link for fork PRs from accounts without write access, so the `Closes #71` keyword in the commit and here does the closing on merge.


mguttmann:
Follow-up pushed (f43ef74): switched-on chips are now always visible. The first version hid the Tasks chip while the task list was empty and offered MCP and Tasks only while the agent process was alive, so a chip someone had just switched on was nowhere to be found. Now Tasks shows its label and an empty state, MCP and Tasks are disabled (dimmed, not openable, no request sent, no badge) while no agent process runs, and the account chip stays disabled during a sign-in instead of vanishing. Only chips that do not exist for the agent kind stay absent. Settings hint and manual updated; 1022 tests, lint and build green.

mguttmann:
Rebased onto v0.2.1 (68a4bc2). The only conflict was the toolbar block in ChatPane.tsx, where v0.2.1 gave the permission chip its pending mark, the picked-mode label and the four-way tooltip; those are now carried inside the chip list's permission entry, so the chip renders exactly as in v0.2.1. Lint, type check and 1035 tests green on the rebased branch.

mguttmann:
Rebased onto v0.2.2 (7874081) and squashed into one commit, because the toolbar block moved under v0.2.2's conversation view work and a two-step replay of the same region produced no reviewable history. Conflicts resolved: the chip list replaces the fixed primary/secondary split as before, the model chip follows your `footer` to `advancedFooter` rename, and termStore keeps both new imports. pnpm test 120 files / 1053 tests, build and lint green.

Issue #73, opened 2026-09-13
Mobile app: native dialogs and errors of the remote plugin are hardcoded Chinese, and the trust button is styled as destructive


mguttmann:
## Summary

The native side of the mobile app still speaks Chinese only. The start page is translatable (it goes through the shared i18n), but the dialogs and messages produced by the Capacitor remote plugin are hardcoded: 66 Chinese strings in `VelaRemotePlugin.swift`, 8 in `URLScanner.swift`, 62 in the Android `VelaRemotePlugin.kt`, and 4 in the shared `download.js`. A German or English user meets them at the most sensitive moments: the certificate and SSH host key fingerprint prompts, the connection error pages, camera permission errors, and the account flow.

Two examples from a real device:

- The fingerprint prompt: title 确认远端指纹, body ending in 请与主机管理员核对。, buttons 取消 and 确认并信任. A user who cannot read it has to guess which button trusts the host. The confirm action is also styled `.destructive` (red), which on iOS signals danger, so the safe guess is the wrong one.
- The failure page after a load error: heading in German (it comes from `MobileText`), body 无法加载远端页面，请检查网络后重试，或返回连接列表。 in Chinese (hardcoded), buttons in German again.

## The mechanism already exists

The plugin has a localization path: `sync-resources.mjs` writes `native-text.json` from the shared dictionaries for a fixed list of nine keys, and `MobileText.swift` / `MobileText.kt` read it at runtime (`MobileText.get("mobile.loadSlow")`). Only those nine strings use it; the other 140 bypass it.

## Proposed fix

1. Move every user-facing native string into `mobile.*` keys in `src/i18n/locales/en.ts` (English written as UI copy, German translated, Simplified Chinese kept verbatim, Traditional Chinese converted, the other locales with the repo's `// TODO translate` marker), and extend the key list in `sync-resources.mjs` so they reach `native-text.json`.
2. Replace the hardcoded strings in Swift, Kotlin and `download.js` with `MobileText.get(...)` (or the JS equivalent for `download.js`).
3. Give the fingerprint confirm action the `.default` style; cancel keeps `.cancel`. Trusting a verified fingerprint is the expected action, not a destructive one.
4. Error strings passed to `call.reject(...)` are shown by the start page as-is, so they should be localized on the native side too, or carry a key the page can translate.

[...]

PR #77, opened 2026-09-13
fix(mobile-ios): show one fingerprint prompt at a time so a URL connection with a self-signed certificate completes (#72)

### LINKED ISSUE #72: iOS app: concurrent TLS trust prompts deadlock a URL connection, so a self-signed desktop can never be reached
## Summary

On a real iPhone, a URL connection to a VelaTerm desktop with a self-signed certificate never completes: the app shows the fingerprint prompt (sometimes twice), then "loading is taking longer", and finally the native "Connection unavailable" page, regardless of whether the user confirms or cancels. Safari on the same phone connects to the same pairing link without trouble, and the desktop log shows a full session from Safari but not a single WebSocket handshake from the app.

Observed with `apps/mobile` at v0.2.0 (`af9e7de`), iOS 27 on an iPhone 17 Pro, desktop VelaTerm 0.2.0 reached over Tailscale (`https://100.117.162.60:8799/#pair=...`).

## Root cause (from the device log)

The iOS remote plugin presents one `UIAlertController` per TLS challenge. `WKWebView` opens several connections while loading the page (document plus resources), each connection raises its own server-trust challenge, and `webView(_:didReceive:completionHandler:)` calls `onCertificate` for every one of them (`VelaRemotePlugin.swift:785-795`). `onCertificate` calls `approve()` (`:535-541`), and `approve()` builds and presents a fresh alert for each call (`:345-356`). There is no coalescing of concurrent challenges for the same identity.

UIKit refuses the second alert while the first is still up. The device log records it as an error at the moment of the first attempt:

```
UIKitCore [com.apple.UIKit:Presentation] Attempt to present <UIAlertController: 0x13070e400>
on <VelaRemotePlugin.ProjectBrowser: 0x1317fca00> which is already presenting <UIAlertController: 0x13070de00>.
```

When `present` fails, the continuation inside `approve()` is never resumed, so that challenge's `completionHandler` is never called. The connection stays pending forever, the page never finishes loading, `pageReady` never fires, and the 30 second `loadSlow` overlay appears. Confirming the visible alert does not help because a different connection is the one that hangs. Cancelling the visible alert produces `NSURLErrorUserCancelledAuthentication` on the other connection, which `failedNavigation` renders as "Connection unavailable".

A related instance of the same class: the first alert is presented on `self.browser`, which was assigned and presented moments earlier (`:543-545`). A challenge that arrives while the browser controller is still animating in hits a view controller that is not yet in the window hierarchy, with the same silent hang.

## What is not the cause

I verified the rest of the chain to be sure the fix is in the right place:

- The pairing fragment survives the QR scan, the saved connection and `establish()` (`safeURL` keeps the fragment, `:121-125`, `:385`).
- Loading the same URL in a `WKWebView` on macOS (same WebKit generation) with the pinned trust and the app's injected globals renders the password page within a second, and the WebSocket challenge is routed through the same delegate and succeeds, so neither the SPA nor the WebSocket trust path is at fault on current WebKit.
- The server certificate's SAN does not list the Tailscale address (it was generated with the addresses of the time), but the trust override in the delegate ignores hostname mismatch, and Safari connected to the same certificate. Regenerating the certificate when the address set changes is worth a separate issue, not this one.

## Proposed fix

1. **Coalesce trust decisions per identity.** Keep one pending decision per `identity` (host and port); concurrent challenges for the same identity await the same result instead of presenting their own alert. After the user decides, every waiting completion handler receives the same answer. The SSH host key path (`:389-394`) shares `approve()` and benefits from the same change.
2. **Present on the top-most view controller that is actually on screen**, walking `presentedViewController` from the bridge's controller, and defer presentation until the browser controller's presentation has completed.
3. **Never leave a continuation unresumed.** If presenting the alert fails for any reason, resume with `false` so the challenge is cancelled and the failure is visible, rather than hanging.
4. **Surface the real error.** `failedNavigation` should show the `NSError` domain and code (and a short explanation for the common ones) instead of a generic text, and the `loadSlow` overlay should say what it is waiting for. That would have made this a five minute diagnosis.

The Android plugin builds one `AlertDialog` per challenge as well (`VelaRemotePlugin.kt:271-273`); Android allows stacked dialogs, so it asks twice rather than hanging, but the same coalescing applies.

I have a fix in progress for iOS and will link the PR here.

mguttmann:
Closes #72
Refs #73 (item 3, the confirm button style)

## What happens today

Details and the device log are in #72. In short: `WKWebView` raises several server-trust challenges while loading the page, the plugin presents one `UIAlertController` per challenge, UIKit refuses the second one while the first is up ("Attempt to present ... which is already presenting"), the refused continuation is never resumed, that connection stays pending, `pageReady` never fires, and the user ends up on "loading takes longer" and then "connection unavailable" no matter which button they tap. Safari on the same phone connects to the same pairing link without trouble.

## What changes

- **`TrustPromptCoordinator.swift` (new, UIKit-free).** Serializes fingerprint prompts: concurrent requests for the same identity and fingerprint share one alert and receive the same decision, distinct requests are queued and shown one after another, every caller is answered exactly once, a presenter failure answers `false` to all of its waiters, and a caller whose connection generation moved on gets `false` without a prompt. Both the TLS path (`onCertificate`) and the SSH host key path (`HostValidator`) go through it because both call `approve()`.
- **`presentTrustAlert`.** Presents on the top-most view controller that is fully in the window hierarchy (walks `presentedViewController`, skips alerts, waits up to about 5 s while the browser is still animating in). A watchdog answers `false` when the alert never appears within 5 s or disappears without an action for 1 s, so no challenge can stay pending.
- **`failedNavigation`** appends the system error text plus `NSError` domain and code to the failure page, so the real cause is visible instead of a generic sentence only.
- **Confirm action styled `.default`** instead of `.destructive`; cancel keeps `.cancel`. Trusting a verified fingerprint is the expected action, not a destructive one.
- **`scripts/test-ios-native.sh`** compiles the coordinator with plain `swiftc` on macOS together with `scripts/ios-native-tests/main.swift`, no device or simulator needed: coalescing, no overlap, throwing presenter, stale generation, a changed fingerprint queued behind an open prompt, and a late joiner. The mobile README documents the script.

Vault semantics are unchanged: a stored matching fingerprint is trusted without a prompt, a `true` decision is persisted once, `false` is never persisted.

## Not in this PR

- Android: `VelaRemotePlugin.kt` builds one `AlertDialog` per challenge too; Android stacks dialogs, so it asks twice rather than hanging. The same coalescing would apply, but I cannot build the Android project here.
- The hardcoded Chinese strings of the native plugin (#73). A follow-up PR moves them into the shared dictionary.

## Verification

- `sh scripts/test-ios-native.sh`: all cases pass.
- iOS simulator build: BUILD SUCCEEDED.
- Real device (iPhone 17 Pro, iOS 27, desktop VelaTerm 0.2.0 with a self-signed certificate over Tailscale): pending at the time of opening, I will report the result here.

Note on linking: this PR comes from a fork without write access, so GitHub does not add it to the issue's Development sidebar on its own; `Closes #72` here and in the commit closes the issue on merge.


mguttmann:
Verified end to end in the iOS simulator (iPhone 17 Pro, iOS 27) against a VelaTerm 0.2.0 desktop with its self-signed certificate, through an opt-in hosted XCTest that drives the real app (`AppTests/TrustPromptE2ETests`, added in #78, skipped unless `VELA_E2E_URL` is set): exactly one fingerprint prompt while WebKit's parallel connections are pending; "Trust and continue" loads the page and persists the fingerprint once; "Cancel" shows the failure page at once and no further prompt follows.

Two follow-ups that this test surfaced landed in #78 rather than here, to keep this PR to the deadlock: after a refusal the page now explains itself (some WebKit builds report the failed load as a plain cancellation, which the failure handler ignores by design, so the page sat on "loading" for 30 s), and declined fingerprints are remembered for the current load, because WebKit opens further connections after a refusal and each of them asked again until the user retries.

On my device the first attempt with this fix ended in `NSURLErrorDomain -1200` right after tapping the bold Cancel action, which is the refusal path working as intended; #78 makes the trust action the highlighted one and translates the labels, so that mistake is harder to make.

mguttmann:
Second root cause found and fixed in this PR (commit dbbbaf9). On a real device the connection still failed after the deadlock fix, with `NSURLErrorDomain -1200` and no prompt at all. The device log shows the TLS flow to the desktop's Tailscale address cancelled in about 16 ms with an ATS violation, before the trust challenge is delivered:

```
boringssl ... Trust evaluate failure: [leaf AnchorTrusted SSLHostname]
boringssl ... Error [ATS Violation]
nw_connection ... reporting state cancelled error ATS violation
Task ... finished with error [-1200] NSURLErrorDomain
```

The app declared only `NSAllowsLocalNetworking`, which relaxes App Transport Security for local hostnames but not for a routable IP (LAN or Tailscale), so iOS refused the self-signed certificate outright and never consulted the plugin's fingerprint pinning. Safari connects because it is not bound by the app's transport policy, and my earlier simulator run passed only because `127.0.0.1` is ATS-exempt.

Fix: `NSAllowsArbitraryLoadsInWebContent` in the app Info.plist. It lifts ATS for `WKWebView` content only, so the plugin's SHA-256 fingerprint pinning in the navigation delegate becomes the trust decision, exactly as it already did over loopback. The native account API to velaterm.com keeps full ATS, because the key is scoped to web content.

Verified in the iOS simulator against the real desktop over its Tailscale IP (not loopback): before the key the trust path times out, after it the page loads in about 4 s and the fingerprint is stored once. The opt-in `AppTests/TrustPromptE2ETests` in #78 covers both.

PR #81, opened 2026-09-15
feat(panes): open existing sessions in splits, tile them, and mark the active split in the sidebar

heywood8:
## What

A split pane could only ever hold a new scratch terminal, so there was no way to put an existing agent session next to another one. This PR lets existing sessions be arranged in splits without restarting them.

- **Sidebar menu:** Open in Split Right, Open in Split Down and Open in Focused Pane, acting on the focused pane of the active tab.
- **Drop onto a pane:** dragging a session from the sidebar splits toward the nearest edge, or replaces the pane's session when dropped in the middle, with a preview of where it lands.
- **Tile Selected Sessions:** two to four sessions open as one evenly divided, pinned tab, from the multi-selection menu or by dragging them onto the stage.
- **Sidebar marker:** sessions in the other panes of the active tab get a faint version of the focused row's highlight.

## How a move stays alive

Unmounting a `TerminalView` kills its process on desktop, and `CenterPane` renders panes keyed by session ID. A moving session therefore leaves its old tree and enters its new one in the same store update: the placements are pure transforms in `src/store/paneMoves.ts`, committed with one `set`.

Tabs are keyed by a session ID and labelled with that session. When a move takes the keying session out of a tab that keeps other panes, the tab is re-keyed to a remaining session; otherwise reopening the departed session would write over the surviving tree and kill every process in it. A named session displaced from a pane moves to a background tab; a scratch terminal closes.

## Scope and non-goals

- No version bump, no changelog entry. New strings are in all 11 locales; only English and Russian were reviewed by a speaker.
- Split trace gains `sidebar`, `drop` and `tile` sources on the frontend and backend.
- Browser nodes stay out of splits: their native WebView draws above the DOM and would hide menus and previews of neighbouring panes.

## Known minor points left open

- `splitNew` captures the active tab before `await getSessionCwd`; switching tabs during that await leaves focus in a hidden tab. Pre-existing.
- `closePane` does not re-key a tab whose keying session it removes, so the tab label can still name a closed session. Pre-existing.
- In one dev session split-trace entries stopped reaching the runtime log after a tile, and one drop did nothing until the window reloaded. Neither has reproduced.

## Tests

New `paneMoves.test.ts`, `paneDrop.test.ts` and `termStore.paneMoves.test.ts` cover the layout transforms, drop geometry, and moves through the real store without killing processes; `split_trace` gains a test for every frontend source.

Rebased onto `7874081` (v0.2.2); the only conflict was an import line in `sessionMenu.tsx`. Verified on the rebased branch: `pnpm test` 1058 passed in 120 files, `tsc --noEmit` and `pnpm lint` clean, `cargo test split_trace` passed. The full `cargo test` suite was not run.

Not verified end to end: Windows, Linux, the Electron shell, remote browser clients and mirror mode. Manually tried in `pnpm dev:desktop` on macOS: drops onto pane edges and centres, tiling, dropping again after closing panes, and the sidebar marker.


heywood8:
image

